PR #573, opened 2015-04-25
Updated some missing part in doc

yangchenyun:


npostavs:
The `.yas-skip.el` was already documented in the mean time, and I've just added some documentation for `.yas-setup.el`.

PR #577, opened 2015-05-14
chmod -x doc/yas-doc-helper.el

barak:
yak shaving


npostavs:
hmm, I wonder it got executable in the first place...

Issue #492, opened 2014-07-18
`yas--modes-to-activate' doesn't work well if the parent mode is an alias


ccyip:
`lua-mode`, for example, is derived from `lua--prog-mode`, which is an alias for `prog-mode` (or `fundamental-mode`, depends on the existence of `prog-mode`). In this case, `yas--modes-to-activate` misses `prog-mode`. As a result, we can't use snippets from `prog-mode` in `lua-mode`.

_Reproduction_

``` emacs-lisp
(let ((major-mode 'lua-mode))
  (yas--modes-to-activate))
;; -> (lua-mode lua--prog-mode)
```

_Possible solution_
Find the end of the chain of aliases of mode.

``` emacs-lisp
;; The following code is copied from `help-fns.el'.
(defun yas--indirect-mode (mode)
  (let ((f mode))
    (while (and (fboundp f)
                (symbolp (symbol-function f)))
      (setq f (symbol-function f)))
    f))

(defun yas--modes-to-activate ()
  "Compute list of mode symbols that are active for `yas-expand'
and friends."
  (let (dfs)
    (setq dfs (lambda (mode &optional explored)
                (push mode explored)
                (cons mode
                      (loop for neighbour
                            in (remove nil (cons
                                            (yas--indirect-mode (get mode
                                                                     'derived-mode-parent))
                                                 (gethash mode yas--parents)))

                            unless (memq neighbour explored)
                            append (funcall dfs neighbour explored)))))
    (remove-duplicates (append yas--extra-modes
                               (funcall dfs major-mode)))))

```

Now we have `(lua-mode prog-mode)`.
Or maybe all aliases should be added to the result as well, then we have `(lua-mode lua--prog-mode prog-mode)`.

PR #604, opened 2015-09-07
remove useless line

huangguiyang:


npostavs:
Removing this line causes the tests to fail, which suggests it's not useless. And you really should provide some justification that something is useless when proposing to remove it.

By using `git blame` you can see when and why this was added in the first place: 0b7b34a333ff5fbb279f57980d6cc983e9a5d8c9, #492, #494

huangguiyang:
I see, sorry ;-)

PR #632, opened 2015-11-26
Use `user-emacs-directory' variable instead of `.emacs.d'

rakkesh:
* yasnippet.el: replace `.emacs.d` with `user-emacs-directory`


rakkesh:
This would be helpful when using custom emacs directories, "~/.emacs-24.5.1", "~/.emacs-25.0.50.1", and etc

rakkesh:
Eh, the tests failed!

PR #645, opened 2016-01-10
Decide field clearing commands based on delsel

npostavs:
It seems that it's not working for `yank` though, somehow when the overlay modification is triggered `this-command` has been set to `t`. Still need to investigate why.

```
* yasnippet.el (yas--skip-and-clear-field-p): Check `delete-selection'
  of `this-command'.
```

### DIFF
--- a/yasnippet-tests.el
+++ b/yasnippet-tests.el
@@ -142,6 +142,39 @@
 ;;     (should (string= (yas--buffer-contents)
 ;;                      "brother from another mother!"))))
 
+(ert-deftest dont-clear-on-partial-deletion-issue-515 ()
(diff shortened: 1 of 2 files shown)


npostavs:
> It seems that it's not working for yank though, somehow when the overlay modification is triggered this-command has been set to t. Still need to investigate why.

Answer: `yank` sets `this-command` to `t` while executing:
```elisp
(defun yank (&optional arg)
  ...
  ;; If we don't get all the way thru, make last-command indicate that
  ;; for the following command.
  (setq this-command t)
  ...
  ;; If we do get all the way thru, make this-command indicate that.
  (if (eq this-command t)
      (setq this-command 'yank))
  ...)
```

PR #675, opened 2016-04-09
use delete-region instead of kill-region in yas--create-snippet-xrefs

bsdf:
i ran into issue #550, and found that for me it was a `pbcopy.el` interaction issue. `pbcopy.el` doesn't seem to like the the many calls to kill-region in `yas--create-snippet-xrefs`. since we don't want a bunch of `\\snippet`s in our kill-ring anyways, we should use `delete-region` instead.

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -4374,8 +4374,8 @@ object satisfying `yas--field-p' to restrict the expansion to.")))
                                          'yasnippet)))
         (when template
           (help-xref-button 1 'help-snippet-def template)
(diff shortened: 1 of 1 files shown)


npostavs:
> since we don't want a bunch of `\\snippets` in our kill-ring anyways, we should use delete-region instead.

Agreed.

Have you signed copyright papers for Emacs? (we can still accept up to 15 lines in changes if you haven't, so this change will still go in regardless)

bsdf:
i have not, but will look into doing so if i have any contributions in the future

PR #676, opened 2016-04-09
Avoid potential problems with user's hooks

### LINKED ISSUE #673: Weird issues since cleanup commit 18f7b1
Hey guys,

Bear with me since I have no repro steps but I will try to explain the symptoms clearly enough for you to find a solution.

First I'm pretty sure it does not come from Spacemacs config but OTOH if I `emacs -Q` and require yasnippet the symptoms does not appear.

Here are the symptoms:
- Starting Emacs gives `(void-function yas--load-pending-jits)`
- So I tried with `emacs -Q` --> works
- Then I removed some configuration of Yasnippet from Spacemacs, still the error
- The error seems to be triggered as soon as I add `yas-minor-mode` to `emacs-lisp-mode-hook`
- I dug into `yasnippet.el` and move around some code, namely moving `yas--load-pending-jits` before `define-minor-mode` and the error disappeared and was replaced by another missing symbol used in `yas--load-pending-jits`
- This was pulling an endless string so I put the definition of the minor mode at the end of the file and then I got a `Recursive load` error.

Sorry for lacking a repro step, I will try to add one at some point. I hope this description of the symptoms will be enough and most importantly I hope to learn why those errors are triggered.

cc @monnier @npostavs

EDIT: these errors let a dangling hook in post-command-hook and prevent Emacs from being usable.

npostavs:
```
* yasnippet.el (yas--font-lock-keywords): ignore-errors from
emacs-lisp-mode call, in case user's hooks cause problems (e.g. enabling
`yas-minor-mode' before we finish loading).
```

fix #673, replace/ close #674.

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -880,7 +880,7 @@ Honour `yas-dont-activate', which see."
 (defvar yas--font-lock-keywords
   (append '(("^#.*$" . font-lock-comment-face))
           (with-temp-buffer
-            (emacs-lisp-mode)
+            (ignore-errors (emacs-lisp-mode))
             (font-lock-set-defaults)
             (if (eq t (car-safe font-lock-keywords))
                 ;; They're "compiled", so extract the source.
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you :+1:

PR #688, opened 2016-05-03
Separate parsing from evaluation of backquote lisp

npostavs:
```
* yasnippet.el (yas--save-backquotes): Parse all backquoted lisp forms
and only then evaluate them.
```

cf #141, but generalizing to multiple expressions

> Problem lies in yas/replace-backquotes function.
It starts with snippet content inserted into buffer. It evaluates as follows:
- parse lisp code
- evaluate lisp code
- remove snippet lisp code from buffer

> It will work if order would be:
- parse lisp code
- remove snippet lisp code from buffer
- evaluate lisp code


npostavs:
For future reference, this change was reverted for compatibility with snippets that insert text by side effect, see #710 (ee4efdbbde2da236fa06bb18290339d0cdca6699). At some point it may be reinstated.

PR #696, opened 2016-05-14
Add yas-backport-obsolete-alias option

zonuexe:
## Problem

Many functions and variables have both prefix `yas-` and `yas/`.  This is useful for users who want to update from the old version, but it is unnecessary feature for users of the latest version.

I am using `helm-M-x`.

## Proposal

Add `yas-backport-obsolete-alias` custom variable.  Don't make alias If this variable is `nil`.

I am not good at English, I think there is a better variable name.

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -386,6 +386,11 @@ the trigger key itself."
   :type '(repeat function)
   :group 'yasnippet)
 
+(defcustom yas-backport-obsolete-alias t
+  "If non-nil backport function and variables from old version of yasnippet."
+  :type 'boolean
+  :group 'yasnippet)
+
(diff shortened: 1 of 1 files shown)


npostavs:
> I think there is a better variable name.

I'm thinking to call it `yas-alias-to-yas/prefix-p`, see #699.

zonuexe:
My copyright assignment paper seems to have been accepted in the FSF.

npostavs:
Ah, that's great, now we don't have to worry about hitting the change size limit.

---

Do you think it's worth adding some code so that the aliases can be removed/added after loading as well?

zonuexe:
move to #699

PR #709, opened 2016-06-13
Yas selected text escape

rubikitch:


npostavs:
Why only escaping for new snippets? Perhaps we should always be escaping the output of backquotes?

npostavs:
Oh I get it, we want to escape it when writing the new snippet body, because escapes will be interpreted when the snippet is expanded.

npostavs:
Do you have copyright assignment? (change is small enough to install regardless)

rubikitch:
yes!

npostavs:
Thanks, I've squashed this to `master`.

PR #718, opened 2016-07-09
Simplify `yas-new-snippet' load&save logic

npostavs:
Slight tweak of code mentioned at https://github.com/joaotavora/yasnippet/pull/694#issuecomment-220967384, I don't want to lose the default file name in `C-c C-c`.

Close #694.


npostavs:
Added an `after-save-hook` to load the snippet. While doing so I noticed `yas--message` had an off-by-1 error, so I added a fix for that.

PR #737, opened 2016-10-07
Replaced all occurances of `capitaomorte'

dotemacs:
Since the new repo destination is now under **joaotavora**, I've fixed all
the stale links to point to the new destination

Also, see this PR: https://github.com/melpa/melpa/pull/4301

Close #704.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,4 +26,4 @@ there is no separate Changelog file.
 For trivial changes, a message consisting of just the changelog entry
 (the `* foo.el ...` part) is fine.
 
-[bugnote]: https://github.com/capitaomorte/yasnippet#important-note-regarding-bug-reporting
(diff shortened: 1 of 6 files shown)


npostavs:
See also #704, there was some thought about moving to yasnippet/yasnippet, but maybe it's not worth the trouble.

dotemacs:
@npostavs thanks for the link to the PR #704 

I don't see any conflicting info in that PR about my changes.

It might be a good idea to have **yasnippet** organisation on GitHub, with **yasnippet/yasnippet** repo.  As it might bring in all the related snippets repos under 'one roof'. But that's outside the scope of this PR and for @joaotavora to decide.

Thanks

npostavs:
Do you have copyright assignment for Emacs? (obviously a trivial search/replace can be considered a small change that can be merged regardless, I'm just asking to know if I need to mark it)

dotemacs:
> Do you have copyright assignment for Emacs? 

No, but if you do, can you do it please? Thanks

PR #757, opened 2016-12-10
No timestamp in HTML head

spwhitton:
Makes the documentation build reproducible.  Thanks.


npostavs:
Thanks, do you have copyright assignment for Emacs? (I'm just asking to know if I should mark it as a small, copyright exempt, change or not).

spwhitton:
No, I don't.  It'll have to be copyright exempt.

-- 
Sean Whitton

npostavs:
I've pushed another commit to add the source commit's timestamp to the HTML output, looks ok?

spwhitton:
The Debian package build process cannot invoke git, so I would have to add a patch to revert f4d19f9, if this were merged in its present condition.

However, we have an environment variable [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) which contains a UNIX timestamp representing the latest change to the source.  Could you check for that environment variable first, and use it in preference to git, if it is set?

npostavs:
Oh, I didn't mean to drop the `yas--version` fallback.

> Could you check for that environment variable first, and use it in preference to git, if it is set?

The `ignore-errors` around the git call should be okay too, right?

spwhitton:
Thank you for your interest in finding a solution that works for both upstream and Debian!

1. `SOURCE_DATE_EPOCH` is a unix timestamp, i.e., the number of seconds since New Year 1970.  It doesn't look like your current commit handles that properly, unless I'm missing something.

2. Would it be possible to check for `SOURCE_DATE_EPOCH` first, and fallback to querying git (at present it's the other way around)?  This would mean that the package will build identically whether or not git is present.  And it won't affect non-Debian users, since they won't have this environment variable set.

Note that `SOURCE_DATE_EPOCH` is taken from the Debian changelog, so it won't be identical to the timestamp of the last git commit.

npostavs:
> 1. `SOURCE_DATE_EPOCH` is a unix timestamp, i.e., the number of seconds since New Year 1970.  It doesn't look like your current commit handles that properly, unless I'm missing something.

Yeah, it wasn't formatting into something readable, fixed now.

> 2. Would it be possible to check for `SOURCE_DATE_EPOCH` first, and fallback to querying git (at present it's the other way around)?  This would mean that the package will build identically whether or not git is present.

Ah right, that makes sense.

> Note that `SOURCE_DATE_EPOCH` is taken from the Debian changelog, so it won't be identical to the timestamp of the last git commit.

An approximate date should be fine.

spwhitton:
Wrong number of arguments: current-time-string, 2

spwhitton:
The patch works fine with emacs25, and the build is reproducible.  You might not think it's worth fixing, but the build fails with emacs24:

       Wrong number of arguments: current-time-string, 2

npostavs:
> You might not think it's worth fixing, but the build fails with emacs24:

Well, it's not too hard to fix.

spwhitton:
LGTM.  Looking forward to seeing a release including this change!  Thanks again.

npostavs:
Thanks for your feedback!

PR #779, opened 2017-02-17
Fix running hook variable set as buffer-local

notetiene:
```
* (yas--safely-call-fun): Check that `fun` isn’t `t`. `t` is added to
  a hook list when it is set as buffer-local when setting the `local`
  argument of `add-hook`.
```


notetiene:
Oh, thanks for the quick response!

I thought about that, but since `yas--safely-call-fun` is just called for hooks and also contains the message "a hook", I found it more complicated to include it in `yas--safely-run-hook` than anything else.

What do you mean by ignoring functions? As far I as I know it just ignores t as a function. Am I right?

npostavs:
> What do you mean by ignoring functions? As far I as I know it just ignores t as a function. Am I right?

If you do
```elisp
(add-hook 'yas-after-exit-snippet-hook 'my-global-hook)
(add-hook 'yas-after-exit-snippet-hook 'my-local-hook nil t)
```
then you have
```
yas-after-exit-snippet-hook’s value is (my-local-hook t)
Local in buffer *scratch*; global value is (my-global-hook)
```

When in `*scratch*`, you would expect both `my-global-hook` and `my-local-hook` to run. The `t` stands for the global value, `(my-global-hook)` in this case, so it should not be ignored.

notetiene:
But in that case, `my-global-hook` is never run even when `t` is present.

notetiene:
The way `add-to-list` works is that is overrides the global hooks when setting a new hook as local.

npostavs:
> But in that case, `my-global-hook` is never run even when `t` is present.

Exactly, that's the bug.

`(elisp) Running Hooks`:
```
 -- Function: run-hooks &rest hookvars
     [...]
     If the hook variable is buffer-local, the buffer-local variable
     will be used instead of the global variable.  However, if the
     buffer-local variable contains the element ‘t’, the global hook
     variable will be run as well.
```

notetiene:
In that case, you're right. I didn't notice that yesterday when reading the C source. There should be a test to also take global values into account. Let me see if global value are run _after_ or _before_ local ones and I'll update the PR.

yasnippet.el
```diff
@@ -1352,7 +1352,7 @@ Returns (TEMPLATES START END). This function respects
       ((debug error) (cdr oops)))))
 
 (defun yas--eval-for-effect (form)
-  (yas--safely-run-hook (apply-partially #'eval form)))
```
npostavs: We might need to bind `debug-on-error` based `yas-good-grace` similar to what `yas--safely-run-hook` does.

yasnippet.el
```diff
@@ -3343,20 +3343,24 @@ This renders the snippet as ordinary text."
...
+  (let ((hook (symbol-value hook-symbol))
+        (debug-on-error (and (not (memq yas-good-grace '(t hooks)))
                              debug-on-error)))
     (if (functionp hook) (yas--safely-call-fun hook)
-      (mapc #'yas--safely-call-fun hook))))
+      (if (not (member t hook))
+          (mapc #'yas--safely-call-fun hook)
+        (setq hook (remove t hook))
+        (mapc #'yas--safely-call-fun hook)
+        (mapc #'yas--safely-call-fun (default-value hook-symbol))))))
```
npostavs: I would prefer to run the `default-value` based on where `t` is located in the list, rather than unconditionally doing it at the end. Although it's true `add-hook` always leaves `t` at the end of the list so we can probably get away with this.

yasnippet.el
```diff
@@ -3367,7 +3371,6 @@ If so cleans up the whole snippet up."
                      (not (and active-field (yas--field-contains-point-p active-field))))
                  (setq snippets-left (delete snippet snippets-left))
                  (setf (yas--snippet-force-exit snippet) nil)
-                 (setq snippet-exit-hook yas-after-exit-snippet-hook)
```
npostavs: This drops the code that gets the snippet-local binding of this hook (thus the failing test). Consider keeping the `snippet-exit-hook` variable and then let-binding `yas-after-exit-snippet-hook` to its value around the `yas--safely-run-hook` instead.

npostavs:
By the way do you have copyright assignment for Emacs? (I think the change is small enough to install regardless, but it's getting close to the edge (15 lines)).

To start the copyright assignment process, fill in this [form](http://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright/request-assign.program), with "Emacs" as the program name, and send it to the mentioned email address.

notetiene:
Thanks for the info, but I already made it :smile:

yasnippet.el
notetiene: Thanks for the info. I was working on resolving that binding error.

notetiene:
I hope this new approach is cleaner and works as expected. I'll write a test later on about locally set hooks.

yasnippet.el
```diff
@@ -3385,7 +3407,10 @@ If so cleans up the whole snippet up."
     (unless (or (null snippets) snippets-left)
       (if snippet-exit-transform
           (yas--eval-for-effect snippet-exit-transform))
-      (yas--safely-run-hook snippet-exit-hook))))
+      (if (eq snippet-exit-hook snippet-exit-hook-env)
```
npostavs: This thing with `...-hook-env` seems a bit over-complicated. Why not just
```elisp
(let ((yas-after-exit-snippet-hook snippet-exit-hook))
  (yas--safe-run-hook 'yas-after-exit-hook))
```

npostavs:
Um, so what about the copyright thing? You're quite a bit over 15 lines by now...

notetiene:
I already signed the copyright assignment for GNU Emacs.

npostavs:
> I already signed the copyright assignment for GNU Emacs.

Great!

---

The more I look at this code, I'm thinking we should just use `run-hook` directly. We're doing it by hand in order to isolate errors to particular hook entries, but it doesn't seem worth the trouble.

notetiene:
> we should just use run-hook

If you were talking about `run-hooks`, absolutely! I didn't even know we could debug hooks with yasnippet. It's a rather uncommon way to do so since it should be the responsibility of the user to handle such things.

Could we just instead use `run-hooks` and wrap a `condition-case` around it?

npostavs:
Yes, right, I meant `run-hooks`.

> Could we just instead use `run-hooks` and wrap a `condition-case` around it?

Yes, exactly. This means an error in one hook function will skip the others, but I think that's perfectly acceptable considering the simplification that brings.

notetiene:
I just took 5h to debug yasnippet entirely to see why tests weren't passing. I finally figured out that I forgot to put a whitespace after the first expanded snippet.

I think everything is now fixed. Thanks for your patience.

notetiene:
@npostavs Anything else that should be done to accept that PR?

yasnippet.el
```diff
@@ -3340,23 +3340,34 @@ This renders the snippet as ordinary text."
       (funcall fun)
     ((debug error)
      (yas--message 2 "Error running %s: %s"
-                   (if (symbolp fun) fun "a hook")
+                   (if (symbolp fun) fun)
```
npostavs: You should remove the `if` entirely and leave just `fun`, it's better to say `Error running (lambda()...)` than `Error running nil`.

yasnippet.el
```diff
@@ -3383,9 +3394,11 @@ If so cleans up the whole snippet up."
                 (t
                  nil)))))
     (unless (or (null snippets) snippets-left)
-      (if snippet-exit-transform
-          (yas--eval-for-effect snippet-exit-transform))
-      (yas--safely-run-hook snippet-exit-hook))))
+      (when snippet-exit-transform
+        (yas--eval-for-effect snippet-exit-transform))
+      (if (not (eq snippet-exit-hook-env yas-after-exit-snippet-hook))
```
npostavs: I believe this `snippet-exit-hook-env` is an unneeded complication.

> It's a way to check that yas-after-exit-hook was overridden in yas--letenv. I didn't know if the snippet had access to the outer scope.

Just keep the `snippet-exit-hook` variable, it gets set to the last snippet's `yas-after-exit-hook` value. Then at the end you can just `(let ((yas-after-exit-snippet-hook snippet-hook)) (yas--safe-run-hooks 'yas-after-exit-snippet-hook))`

yasnippet.el
npostavs: This still applies.

npostavs:
> It just took me 5h to debug yasnippet entirely to see why tests weren't passing.

Hmm, do you think the test framework could be changed somehow to make it easier to find such issues?

yasnippet.el
notetiene: True, I forgot about that.

notetiene:
@npostavs I don't think so. It was a really silly error I should have been aware.

However, I think it's a good practice to document every functions according to `checkdoc`. It's much easier to understand and faster to get accustomed with the package.

I also hope everything you requested is there. I had a hard time understanding your point of `snippet-exit-hook-env` as unneeded! I'm not used at all with Emacs Lisp lexical scoping, but it's time to I guess.

npostavs:
Yup, I think we're good, I'll merge sometime this weekend.

> had a hard time understanding your point of `snippet-exit-hook-env` as unneeded! I'm not used at all with Emacs Lisp lexical scoping, but it's time to I guess.

Actually, this was more about dynamic scoping :), but also the way snippet variables work is its own weird thing.

notetiene:
Dynamic lexical scoping :wink:

npostavs:
I've squashed this and made a few minor changes, see #783.

PR #783, opened 2017-02-27
Fix running hook variable set as buffer-local (v2)

### LINKED ISSUE #780: Setting a yasnippet hook as buffer-local displays an error
When setting a yasnippet hook buffer-local (`yas-after-exit-snippet-hook` or `yas-before-expand-snippet-hook`) `yas--safely-run-hook` calls `yas--safely-call-fun` with `t` as the argument.

`t` is added to a hook variable list when setting a hook as buffer local with `add-hook`. For example:
```elisp
(add-hook 'yas-after-exit-snippet-hook 'my-hook nil t)
```
Would display an error message (and possibly stop execution?) when the `yas-after-exit-snippet-hook` hooks are called.

I've made fix with #779.

npostavs:
This is #779 squashed and tweaked a bit.
Replace/ close #779.
Fix #780.

```
* yasnippet.el (yas--eval-for-effect): Use `yas--safely-call-fun'
instead of `yas--safely-run-hook'.
(yas--safely-call-fun): Display the called argument (even a lambda) on
errors.
(yas--safely-run-hook): Take a hook variable instead of the hook value
itself.  Use `run-hooks' to take buffer-local values into account.
This means it’s no longer possible to warn which hook function had an
error.
(yas--check-commit-snippet): Refactor to conform to
`yas--safely-run-hooks'.
* yasnippet-tests.el (snippet-exit-hooks-bindings): New test that
checks for buffer-local, global and env hooks support.
```

### DIFF
--- a/yasnippet-tests.el
+++ b/yasnippet-tests.el
@@ -583,6 +583,38 @@ SUB"))))
          (ert-simulate-command '(yas-next-field))
          (should (eq yas--ran-exit-hook t)))))))
 
+(ert-deftest snippet-exit-hooks-bindings ()
+  "Check that `yas-after-exit-snippet-hook' is handled correctly
+in the case of a buffer-local variable and being overwritten by
+the expand-env field."
+  (defvar yas--ran-exit-hook)
+  (with-temp-buffer
+    (yas-saving-variables
+     (let ((yas--ran-exit-hook nil)
+           (yas-triggers-in-field t)
+           (yas-after-exit-snippet-hook nil))
+       (yas-with-snippet-dirs
+         '((".emacs.d/snippets"
+            ("emacs-lisp-mode"
+             ("foo" . "foobar\n")
+             ("baz" . "\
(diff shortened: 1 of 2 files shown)


notetiene:
Is there a reason why private functions generally don't have documentation?

I'm asking that because I added a docstring to a function and it was stripped away.

npostavs:
I don't think I removed any, I only shortened to docstring on `yas--safely-run-hook` a bit.

Generally, private functions should have docstrings or at least commentary too, unless it's something trivial, where the Lisp code would be shorter than the English explanation.

notetiene:
Oh, I thought the `yas--safely-run-hook` docstring was gone. Tanks for simplifying it :+1:

It was a general question because I see many private functions don't have a docstring.

npostavs:
> It was a general question because I see many private functions don't have a docstring.

Yeah, I try to add docstrings whenever I'm updating those.

npostavs:
> Tanks for simplifying it

Thanks for the tanks! ;) Also, thanks for reporting and fixing the bug! :D

PR #786, opened 2017-03-08
Fix mirror transformation error with expand-env

notetiene:
* yasnippet.el (yas-expand-snippet): Use the snippet lexical binding
  when moving field. This prevents getting an error when calling a
  function variable defined in expand-env.

* yasnippet-tests.el (snippet-mirror-bindings): Add a test to check if
  we can call a function variable defined in the snippet environment.

Note: I also saw there could be a bug (not related to that patch). If the mirror transformation doesn't contain any character before (at indentation), blank spaces after the mirror will be removed. This is because the mirror transformation may return `""`.

### DIFF
--- a/yasnippet-tests.el
+++ b/yasnippet-tests.el
@@ -614,6 +614,26 @@ foobaz\n"))))
          (should (eq 'global (nth 0 yas--ran-exit-hook)))
          (should (eq 'local (nth 1 yas--ran-exit-hook))))))))
 
+(ert-deftest snippet-mirror-bindings ()
+  "Check that variables defined with the expand-env field are
+accessible from mirror transformations."
+  (with-temp-buffer
+    (yas-saving-variables
+     (let ((yas-triggers-in-field t)
+           (yas-good-grace nil))
+       (yas-with-snippet-dirs
+         '((".emacs.d/snippets"
+            ("emacs-lisp-mode"
+             ("baz" . "\
+# expand-env: ((func #'upcase))
+# --
+hello ${1:$(when (stringp yas-text) (funcall func yas-text))} foo${1:$$(concat \"baz\")}$0"))))
+         (yas-reload-all)
+         (emacs-lisp-mode)
+         (yas-minor-mode +1)
+         (insert "baz")
+         (ert-simulate-command '(yas-expand))
+         (should (string= (yas--buffer-contents) "hello BAZ foobaz\n")))))))
 
 (defvar yas--barbaz)
 (defvar yas--foobarbaz)
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -3731,7 +3731,8 @@ considered when expanding the snippet."
            (let ((first-field (car (yas--snippet-fields snippet))))
              (when first-field
                (sit-for 0) ;; fix issue 125
-               (yas--move-to-field snippet first-field)))
+               (yas--letenv (yas--snippet-expand-env snippet)
+                 (yas--move-to-field snippet first-field))))
            (yas--message 4 "snippet expanded.")
            t))))
 


npostavs:
Thanks. This isn't specific to "function variables" though, right? Any variable binding would behave the same.

notetiene:
No, not specific. However, I don't see how someone would need it in a transformation mirror (except to do a snippet global variable).

notetiene:
For the bug, here's an example:
```
# name: baz
# --
$1 foo ${1:$$(print "bar")}
```
would expand to `barfoo bar` and not `bar`foo bar`.

notetiene:
I just removed the unnecessary `yas-after-exit-snippet-hook`. It was taken from the last test.

npostavs:
Is the `${1:$$(concat \"baz\")}` part of the test from the other bug? I would prefer to deal with that separately.

Also, could you please make the commit message and test docstring describe only the general variable case? I find the talk of "function variable" kind of confusing. It should be enough that the test case uses a variable bound to a function as the example.

notetiene:
Oh sorry, I meant the somehow related bug is:
```
# name: baz
# --
$1 foo ${1:$$(print "bar")}
```

And the actual bug is: `${1:$$(concat "baz")}`. Sure I could change the docs.

notetiene:
Is the new commit okay? I've rebased it to change the commit message and squashed the other commit.

yasnippet-tests.el
```diff
@@ -614,6 +614,26 @@ foobaz\n"))))
          (should (eq 'global (nth 0 yas--ran-exit-hook)))
          (should (eq 'local (nth 1 yas--ran-exit-hook))))))))
 
+(ert-deftest snippet-mirror-bindings ()
+  "Check that variables defined with the expand-env field is
```
npostavs: s/is/are

yasnippet-tests.el
```diff
@@ -614,6 +614,26 @@ foobaz\n"))))
...
+accessible with the mirror transformations."
+  (with-temp-buffer
+    (yas-saving-variables
+     (let ((yas-triggers-in-field t)
+           (yas-good-grace nil))
+       (yas-with-snippet-dirs
+         '((".emacs.d/snippets"
+            ("emacs-lisp-mode"
+             ("baz" . "\
+# expand-env: ((func (lambda (str) (upcase str))))
```
npostavs: Could be simplified as `((func #'upcase))`

yasnippet-tests.el
```diff
@@ -614,6 +614,26 @@ foobaz\n"))))
...
+# --
+hello ${1:$(when (stringp yas-text) (funcall func yas-text))} foo${1:$$(concat \"baz\")}$0"))))
```
npostavs: The latter `${1:$$(concat \"baz\")}` isn't really needed for this test, right?

yasnippet-tests.el
```diff
@@ -614,6 +614,26 @@ foobaz\n"))))
          (should (eq 'global (nth 0 yas--ran-exit-hook)))
          (should (eq 'local (nth 1 yas--ran-exit-hook))))))))
 
+(ert-deftest snippet-mirror-bindings ()
+  "Check that variables defined with the expand-env field is
+accessible with the mirror transformations."
```
npostavs: Should be "accessible from mirror transformations" I think.

yasnippet-tests.el
notetiene: You're right! Forgot about that syntax

yasnippet-tests.el
notetiene: It is. It's really when we use `${n:$$(...)}` that the problem arises. I could have used `print` instead of `concat`.

notetiene:
Thanks for grammar corrections. Next time I should double check.

yasnippet-tests.el
npostavs: Hmm, maybe I've misunderstood what the problem is, couldn't you just do 
```
hello ${1:$$(when (stringp yas-text) (funcall func yas-text))} foo$0
```

PR #803, opened 2017-04-13
Release 0.12.0

npostavs:
0.12.0 should be released soon. Beginning of May perhaps.


zw963:
Cool~  hope #806 could fix in this version.

npostavs:
Yes, it's a regression from 0.11, it will definitely be fixed in 0.12.

zw963:
Still not release 0.12 ? 

current is Jun 17. : )

zw963:
Cool

npostavs:
> Still not release 0.12 ?
> 
> current is Jun 17. : )

It's pushed. Only 2 months late :)

PR #807, opened 2017-04-17
Fix condition timestamps in key filter functions

### LINKED ISSUE #806: 0.11 not worked with some self-defined condition.
Just update to the HEAD vesion of snippets (47f37f7), which broken I some snippets which is worked
in 0.10.

Following is a example:

```cl
# -*- mode: snippet -*-
# name: some snippet
# key: c
# binding: c
# condition: (yas-use-region-p)
# --
foo`(yas-select-text)`bar
```

I expect this snippets will expand in two condition:

1. when I press ``c``, and than ``TAB``
2. Add a mark to select some content in to region, and press ``c``, I expect ``yas-selected-text`` is 
replaced with content.

But, my condition ``(yas-use-region-p)`` is not worked anymore, caused this snippets not expanded.

following is function define:

```el
(defun expand-from-key-p ()
  "通过 TAB 来激活的 snippet."
  (equal (this-command-keys-vector) '[tab]))

(yas-define-condition-cache yas-use-region-p (or (expand-from-key-p) (use-region-p)))
```

Thanks

npostavs:
Fixes #806.

```
* yasnippet.el (yas--maybe-expand-key-filter):
(yas--maybe-expand-from-keymap-filter): Let-bind
`yas--condition-cache-timestamp' to the current time.
```

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -580,7 +580,9 @@ conditions.
   "Holds the YASnippet menu.")
 
 (defun yas--maybe-expand-key-filter (cmd)
-  (if (yas--templates-for-key-at-point) cmd))
+  (when (let ((yas--condition-cache-timestamp (current-time)))
+          (yas--templates-for-key-at-point))
+    cmd))
 
 (defconst yas-maybe-expand
   '(menu-item "" yas-expand :filter yas--maybe-expand-key-filter)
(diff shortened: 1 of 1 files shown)


zw963:
I have confirmed, this worked.

Thanks!

npostavs:
Thanks for testing

PR #875, opened 2017-12-14
Prioritize current major mode in `yas--modes-to-activate`

amosbird:


npostavs:
Can you explain the problem you are trying to solve? As you can see by the test failure, the current order of modes is not there by accident.

amosbird:
Um, I have this `(add-hook! 'yas-minor-mode-hook (yas-activate-extra-mode 'fundamental-mode))` in my init.el and `yas-new-snippet` always tries to save to fundamental-mode directory instead of current buffer's major mode.

npostavs:
Does #887 solve it for you?

PR #873, opened 2017-12-06
Check yas--original-auto-fill-function before using it

chep:
Sometimes when using tramp, I've got an error (Symbol's function definition is
void: nil) because the variable yas--original-auto-fill-function is not
initialized.

A simple check of the value solved my problem.

Issue #912, opened 2018-03-08
wrong indentation when setting is fixed


it6:
I have this setting globally in init```  (setq yas-indent-line 'fixed)``` 

but I see the following behavior, indentation is off in the new release.

![indent-yasnippet](https://user-images.githubusercontent.com/5510191/37135926-6d802ee2-226d-11e8-9329-ebc3eee4bee9.gif)

npostavs:
Oh right, I noticed that bug for `auto` indentation, but it usually doesn't matter in that case because the auto indentation is usually a nop for the first line anyway. Now that I've started using the same machinery for the `fixed` case, it shows up more obviously. Should be fixed in #913.

it6:
can you please merge  #913

PR #920, opened 2018-03-17
; Try to get more info about #873/919

npostavs:
Ref: #873 #919
Instead of asking users one by one to add code to their init, might as well opt-in by default. Plus, give a warning rather than interrupting their other code with an error.
```
* yasnippet.el (yas--watch-autofill): New variable watcher to catch
the problem.
(yas--auto-fill): If `yas--original-auto-fill-function' is nil, print
a warning with some info instead of signaling an error.
```

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -582,6 +582,15 @@ override bindings from other packages (e.g., `company-mode')."
   "The original value of `auto-fill-function'.")
 (make-variable-buffer-local 'yas--original-auto-fill-function)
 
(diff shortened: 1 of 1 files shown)


npostavs:
Turns out calling `debug` during `kill-all-locals-variables` triggers an assertion failure in Emacs (cf [Bug#30846](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=30846)). But we shouldn't trigger for that case anyway (ref #912).

PR #929, opened 2018-04-20
Release 0.13.0

npostavs:
I was hoping to solve #919 first, but it looks like that's unlikely to happen anytime soon.


npostavs:
0.13.0 is released: 5a29fe67d70de3749e30ebb37206395b4d1a51a8

PR #940, opened 2018-05-22
Remove compatibility code for Emacs 22 and earlier

npostavs:
I doubt yasnippet even works on Emacs 22 anyway. The automated tests only go back to 23.4, and I can't even build earlier than 24.3 locally.

```
* yasnippet.el (region-active-p, locate-dominating-stop-dir-regexp)
(locate-dominating-file): Remove.
```

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -618,9 +618,6 @@ override bindings from other packages (e.g., `company-mode')."
 
 ;;; Minor mode stuff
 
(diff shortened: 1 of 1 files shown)


npostavs:
Also removing `last-buffer-undo-list` which is allegedly for "Carbon Emacs". Looking at EmacsWiki, it seems to be something that's pretty obsolete.

https://www.emacswiki.org/emacs/CarbonEmacsPackage

PR #942, opened 2018-06-04
add snippets/ and yasmate/ to .gitignore

truesilver92:
There are files generated through use of this project as a git submodule and package in emacs. It is much nicer to not have them exist.


npostavs:
> There are files generated through use of this project as a git submodule

Hmm, I don't think that's the case, you must have some leftovers from before the submodules were removed.

PR #993, opened 2019-04-14
Allow disabling snippet keybindings from hook

### LINKED ISSUE #987: Filter hook for all commands in yas-keymap?
It would be helpful if there was a hook that we could add predicates to, to determine whether any bindings from this keymap should be active at a given moment.

This should help solve the big remaining problem we have in company-mode WRT key bindings, namely that `TAB` binding for `company-complete-common` cannot work when point is inside a snippet field. The only other approach  for fixing this (AFAIK) is to company-mode to transition to setting the current keymap via overlays as well, which will mean again futzing around with adding/removing newlines, because point must be strictly inside an overlay (and it can't happen when point is at eob).

This comment gave me the idea: https://github.com/company-mode/company-mode/issues/881#issuecomment-477757558

npostavs:
Fixes #987.

```
* yasnippet.el (yas-keymap-disable-hook): New hook.
(yas-filtered-definition): New function.
(yas-keymap): Use it.
```

### DIFF
--- a/doc/faq.org
+++ b/doc/faq.org
@@ -49,21 +49,24 @@ or the [[Github issue tracker][https://github.com/joaotavora/yasnippet/issues/]]
 
 * How do I use alternative keys, i.e. not TAB?
 
-Edit the keymaps [[sym:yas-minor-mode-map][=yas-minor-mode-map=]] and
-[[sym:yas-keymap][=yas-keymap=]] as you would any other keymap:
+Edit the keymaps [[sym:yas-minor-mode-map][=yas-minor-mode-map=]] and [[sym:yas-keymap][=yas-keymap=]] as you would
+any other keymap, but use [[sym:yas-filtered-definition][=yas-filtered-definition=]] on the definition
+if you want to respect [[sym:yas-keymap-disable-hook][=yas-keymap-disable-hook=]]:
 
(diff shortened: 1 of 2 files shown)


doc/faq.org
```diff
@@ -49,21 +49,24 @@ or the [[Github issue tracker][https://github.com/joaotavora/yasnippet/issues/]]
...
+  (define-key yas-minor-mode-map (kbd "TAB") nil)
+  (define-key yas-minor-mode-map (kbd "<the new key>") yas-maybe-expand)
+
+  ;;keys for navigation
+  (define-key yas-keymap [(tab)]       nil)
+  (define-key yas-keymap (kbd "TAB")   nil)
+  (define-key yas-keymap [(shift tab)] nil)
+  (define-key yas-keymap [backtab]     nil)
+  (define-key yas-keymap (kbd "<new-next-field-key>")
+    (yas-keymap-disable-hook 'yas-next-field-or-maybe-expand))
```
dgutov: I think you wanted to use yas-filtered-definition in this example.

npostavs: Yep, thanks. Fixed in 6a738b581f79f11f13b8dd205520c562bd2fa2f5

PR #1002, opened 2019-05-01
Fix field adjustment on deletion

### LINKED ISSUE #996: Can't work with company-mode
Here is a screen record video which shows the problem. http://121.43.160.118:3080/upload/yasnippet_bug2.mp4

Reproduce steps:
1. emacs -Q
2. install these packages from melpa:  yasnippet, company-mode, go-mode, company-go
3. install golang and package gocode: go get -u github.com/mdempsky/gocode
4. open the test file: a.go
5. set `company-backends` to '(company-go)
6. turn on company-mode and yas-minor-mode
7. do as in the screen record video

The test file a.go and the snippets used are uploaded to: http://121.43.160.118:3080/upload/yasnippets_bug2_files.tar.bz2

The order of snippets in the video:
1. sel
2. ch

This problem happens almost everyday for me. But it can't be 100% reproduced before until now.

npostavs:
Fixes #996
```
For deletion, we need to check the bounds before the deletion happens,
otherwise the overlay may already be moved to wrong place.
* yasnippet.el (yas--before-change-modified-snippets): New variable.
(yas--gather-active-snippets): New function.
(yas--on-field-overlay-modification): Use it.
```

### DIFF
--- a/yasnippet-tests.el
+++ b/yasnippet-tests.el
@@ -1122,6 +1122,27 @@ hello ${1:$(when (stringp yas-text) (funcall func yas-text))} foo${1:$$(concat \
      (ert-simulate-command '(yas-next-field-or-maybe-expand))
      (should (string= (buffer-string) "<-<-abcdef\n")))))
 
+(ert-deftest nested-snippet-expansion-5-nested-delete ()
+  "See Github #996."
+  (let ((yas-triggers-in-field t))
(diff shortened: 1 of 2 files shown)


npostavs:
Hmm, for earlier Emacs: ```the function `delete-consecutive-dups' is not known to be defined.```

PR #1003, opened 2019-05-08
Fix FAQ: yas-prev should be yas-prev-field

billsacks:
I got an error when trying to define a keyboard shortcut for yas-prev,
as given in the FAQ. It appears that the correct function definition is
yas-prev-field.

Thank you for this really fantastic package!


npostavs:
> It appears that the correct function definition is yas-prev-field.

Yep, thanks.

PR #1014, opened 2019-07-18
Fix can't insert a new snippet.

baohaojun:
When I use yas-new-snippet to create a new snippet, when the scrippet is
with a complex name (such as "test.snippet"), after I save the snippet,
I can't use yas-insert-snippet to insert it. I do can see it in the menu
bar, though.

This patch fixes this problem.

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -1399,6 +1399,9 @@ conditions to filter out potential expansions."
                                 (push (cons name template) acc))
                             namehash))
                (yas--table-hash table))
+      (maphash #'(lambda (uuid template)
+                   (push (cons uuid template) acc))
+               (yas--table-uuidhash table))
       (yas--filter-templates-by-condition acc))))
 
 (defun yas--templates-for-key-at-point ()


npostavs:
> When I use yas-new-snippet to create a new snippet, when the scrippet
> is with a complex name (such as "test.snippet"), after I save the
> snippet, I can't use yas-insert-snippet to insert it

Can you specify the exact steps you're taking?  Can you reproduce it
from `emacs -Q`?  I tried making a snippet named `test.snippet` and
`yas-insert-snippet` seemed to handle it fine.

baohaojun:
Yes, I can reproduce it using emacs -Q. And sorry for not being clear
the first time, here's the steps:

1.  Create a snippet such as this

    ``` example
    # -*- mode: snippet -*-
    # name: test.snippet.123
    # --
    hello world
    ```

    Note that there is no `key:`. This is very important, if you add a
    `key: test` line, then it won't reproduce: you can use
    yas-insert-snippet.

2.  Try M-x `yas-insert-snippet`, you can't find `test.snippet.123`.

npostavs:
>     1. Note that there is no `key:`. This is very important, if you add a
>        `key: test` line, then it won't reproduce: you can use
>        yas-insert-snippet.

Ah, yes, I can see it now. Your fix seems to be correct, so I just edited the commit message and applied.

PR #1027, opened 2019-11-06
add hook of yas-next-field-hook

ChongQing520:
may i add this hook?

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -3390,6 +3390,7 @@ If there's none, exit the snippet."
          (snippet (car (yas-active-snippets (yas--field-start active-field)
                                             (yas--field-end active-field))))
          (target-field (yas--find-next-field arg snippet active-field)))
+    (run-hook-with-args 'yas-next-field-hook active-field snippet)
     (yas--letenv (yas--snippet-expand-env snippet)
       ;; Apply transform to active field.
       (when active-field


npostavs:
What sort of things do you use this hook for? What about when the user goes back one field, how should hook elements handle that case?

ChongQing520:
1. What sort of things do you use this hook for?

   I only want a default value when user press the <tab>.

2.  What about when the user goes back one field,

   it can restore the previous text , or change to another value.

3.  how should hook elements handle that case?

   something he want  or nothing.

ChongQing520:
``` elisp
(cl-defstruct (yas--field (:constructor yas--make-field (number start end parent-field)))
  "A field.

NUMBER is the field number.
START and END are mostly buffer markers, but see \"apropos markers-to-points\".
PARENT-FIELD is a `yas--field' this field is nested under, or nil.
MIRRORS is a list of `yas--mirror's
TRANSFORM is a lisp form.
MODIFIED-P is a boolean set to true once user inputs text.
NEXT is another `yas--field' or `yas--mirror' or `yas--exit'.
"
  number
  start end
  parent-field
  (mirrors '())
  (transform nil)
  (modified-p nil)
  next
  default  ;; my added
  content
  raw
  )
```

ChongQing520:
``` elisp
(defun chong-yas-next (active-field snippet)
  (when (and (not (yas--field-modified-p active-field))
             (yas--field-default active-field)
             )
    (save-excursion
      (let* ((start (yas--field-start active-field))
             (end (yas--field-end active-field))
             (default (yas--field-default active-field))
             (len (- end start))
             )
        (goto-char start)
        (cond
         ((stringp default)
          (insert (yas--restore-escapes default)))
         ((listp default)
          (insert (yas--eval-for-string default))
          ))))))

(add-hook 'yas-next-field-hook 'chong-yas-next)
```

npostavs:
>     1. I only want a default value when user press the <tab>.

I don't understand, could you say more?

>     2\. What about when the user goes back one field,
>         it can restore the previous text , or change to another value.

I don't want to change the current behaviour here.

>     3\. how should hook elements handle that case?
>         something he want  or nothing.

I meant more, how would a hook function even know that the user is moving backwards?

ChongQing520:
in the function of open

```
fopen(char * fileName , char* mode)
```
i  cannot remember which character can be used as the mode char.

so i can use the snippet

```
fopen(${1:char * filename},${2: r:read  w:write b:binary @"r"@})
```

the "r" is the default value.

ChongQing520:
```
What about when the user goes back one field.
how would a hook function even know that the user is moving backwards?
```
this hook will be called before leaving a yas-field.

move backward or forward is donnot matter.

Issue #1041, opened 2020-03-10
yasnippet breaks iedit-mode


netjune:
When iedit-mode is on, if you expand a snippet by yas-expand-snippet, iedit-mode will be breaked. Because it depends on inhibit-modification-hooks to be nil to run its hook. But yasnippet bind it to t when it creates snippet. So iedit-mode can't work.

npostavs:
It's more complicated than just `inhibit-modification-hooks. `yas--snippet-create` already does `(run-hook-with-args 'before-change-functions begin begin)` and `(run-hook-with-args 'after-change-functions (point-min) (point-max) 0)`, so that part should be covered.

But how should these two packages interact together? I guess if you expand a snippet in one mode then you get a copy in each iedit region, but yasnippet can't handle multiple snippets at a time, so it's not very easy to make this work.

netjune:
In fact it is an empty snippet. It is only a simple string. There are no snippet fields in it. company-lsp always calls yas-expand to try to expand the completion item as a snippet even it is not.
After yas-expand returns, there is no active snippet in buffer.

npostavs:
Oh, the problem is that `inhibit-modification-hooks` blocks overlay
modification hooks too, which we don't call outselves.  Does the patch
in #1046 help?

netjune:
Yes. It works now. Thanks.

netjune:
A related question: Is it safe for yasnippet to abort when there are multiple active snippets? It may happens by accidently.

npostavs:
Thanks for testing.

> Is it safe for yasnippet to abort when there are multiple active snippets? It may happens by accident.

I'm not sure what you mean.

netjune:
When iedit-mode is active, if a snippet with more than one fields is expanded by accident, there will be multiple active snippets and buffer text is messed. Is there a safe way to avoid this?

npostavs:
> When iedit-mode is active, if a snippet with more than one fields is expanded by accident, there will be multiple active snippets and buffer text is messed.

Actually, I think iedit would copy only the text, not the yasnippet field structures, so snippets won't be duplicated.

netjune:
> > When iedit-mode is active, if a snippet with more than one fields is expanded by accident, there will be multiple active snippets and buffer text is messed.
> 
> Actually, I think iedit would copy only the text, not the yasnippet field structures, so snippets won't be duplicated.

It seems that it works sometimes but not always.

netjune:
[...]

PR #1051, opened 2020-05-14
Narrow scope of undo disabling

npostavs:
Might fix additional problem reported in
https://github.com/joaotavora/yasnippet/issues/1041#issuecomment-619466606

```
* yasnippet.el (yas--snippet-create): Don't bind buffer-undo-list
around before and after change-functions.
```


npostavs:
> Might fix additional problem

Turns out it doesn't; closing.

https://github.com/joaotavora/yasnippet/issues/1041#issuecomment-639258122

Issue #767, opened 2017-01-20
automatically avoid extraneous newlines


holocronweaver:
As per #365 and #192, a fairly generic Emacs setup with `mode-require-final-newline` set to 1 causes `snippet-mode` to include extra newlines, much to the chagrin of a new user like myself.

My understanding is the general workaround is to add a mode hook:
`(defun fix-snippet-mode()
    (setq require-final-newline nil))
  (add-hook 'snippet-mode-hook #'fix-snippet-mode)`

But why does snippet mode use `mode-require-final-newline` at all? Wouldn't it be better for `snippet-mode` set `require-final-newline` to `nil` by default and avoid the headache for new users?

holocronweaver:
After looking at yasnippet source, it seems `snippet-mode` does try to set require-final-newline to nil, yet I am still getting extra newlines inserted at the end of my snippets. Is it the expansion of the snippet that is causing the issue, and not the snippet file?

npostavs:
Do the files themselves end with newlines? Can you reproduce this starting from [`emacs -Q`][emacs-Q]?

[emacs-Q]: http://www.lunaryorn.com/2015/11/29/reproduce-bugs-in-emacs-Q.html

holocronweaver:
Figured out what was happening! I was editing some of my snippets in `text-mode` because I accidentally left out the the mode file header. That mode respects `mode-require-final-newline`, which in my case is `t` and hence newlines were being inserted when I saved. Switching to `snippet-mode` did not help since it can't presume trailing newlines are accidental and thus can't trim them.

For anyone experiencing this problem in the future, be sure to only edit your snippets in `snippet-mode`! And if you use another mode, be sure to switch to `snippet-mode` (or locally set `require-final-newline` to `nil`) and then remove the trailing newlines by hand.

I thought I had removed all the trailing newlines, but when I saved in `text-mode` the `require-final-newline` flag was causing them to be added back, doh!

holocronweaver:
Thanks to @npostavs for making me second guess myself about the file endings. That was the key.

npostavs:
Hmm, I wonder if it's worth doing something like:
```elisp
(add-to-list
 'magic-fallback-mode-alist
 `(,(lambda () (when buffer-file-name
		 (member
		  (expand-file-name
		   ".."
		   (file-name-directory buffer-file-name))
		  (yas-snippet-dirs))))
   . snippet-mode))
```
to avoid this kind of thing.

[...]

Issue #365, opened 2013-02-20
Inserting snippets on mid-line selections causes insertion of spurious newline


edgimar:
If using yasnippet when the selected text is only a portion of a line somewhere in the middle of a line, it seems that inserting a snippet causes a newline character to be inserted, even when it might not be desired.  If this is sometimes desirable (when?), maybe this behavior could be made configurable via a configuration option?

To reproduce, put the following in ~/.emacs.d/snippets/latex-mode/bf

```
# name: bf
# key: bf
# --
\textbf{`yas-selected-text`}$0
```

Then run emacs, adding something like the following to a `test.tex` file:

```
This is a test that does NOT work correctly.
```

Highlight `NOT` and run M-x yas-insert-snippet, choosing the 'bf' snippet.

The result will be the following:

```
This is a test that does \textbf{NOT}
 work correctly.
```

The desired result is that there is no newline inserted after the '}' character.  This was tested with a completely empty init.el file except for the few lines necessary to add the yasnippet directory to the path, load yasnippet and enable global mode.

I'm using yasnippet git rev. 2cf544b, and emacs 23.3.1.

PR #1060, opened 2020-06-23
Add yas-snippet-trim feature

conao3:
Hi!
First, thanks for such a awesome package!

I add `snippet-trim` feature and add its optional variable named
`yas-snippet-trim`.

This feature is disabled at default, but if the user enable,
`string-trim` snippet before expand snippet.

This is a great solution for #767, #365, #192.
And this issue is also written in [FAQ](https://github.com/joaotavora/yasnippet/blob/e45e3de357fbd4289fcfa3dd26aaa7be357fb0b8/doc/faq.org#why-is-there-an-extra-newline)

### DIFF
--- a/yasnippet-tests.el
+++ b/yasnippet-tests.el
@@ -1768,6 +1768,20 @@ add the snippets associated with the given mode."
        (yas-should-not-expand '("_"))
        (yas-should-expand '(("car" . "(car )")))))))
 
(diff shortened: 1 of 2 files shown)


conao3:
FYI: @holocronweaver, @edgimar, @joaotavora

conao3:
outdated. close.

PR #1066, opened 2020-08-15
Support using regexp for keys

### LINKED ISSUE #988: Adding support for regex as keys when expanding snippets
I read https://castel.dev/post/lecture-notes-1/ and wanted to replicate at least the simpler parts of the fraction snippet. Is it possible to easily write such snippets in yasnippet where the key is a regex?

In case it is not currently possible I have hacked together some elisp code using the interface provided that can do this but it is not very integrated into yasnippet. I have basically made a new keybinding that checks if a regex has matched. If it matched a corresponding snippet is expanded using `yas-lookup-snippet` and `yas-expand-snippet`. The configuration still happens in elisp code and not in a snippet buffer which is suboptimal.

Is anyone else interested in using regex as keys? If so I could with some help integrate this more into yasnippet and merge it.

Zetagon:
Fixes #988

I'm not sure if the commit messages or the code looks good so I'll mark this as a draft. I have signed the copyright-papers so that should not be an issue.

### DIFF
--- a/yasnippet-tests.el
+++ b/yasnippet-tests.el
@@ -1671,6 +1671,110 @@ add the snippets associated with the given mode."
        (yas-should-expand '(("car" . "(car )")))))))
 
 
+(ert-deftest regexp-key-simple-integration-test ()
+  (clrhash yas--tables)
+  (with-temp-buffer
+    (insert "# name: test\n# regexp-key: [a-z]*\n# --\nfoo")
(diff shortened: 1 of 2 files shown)


SoutrikBandyopadhyay:
Any updates on this ?

SoutrikBandyopadhyay:
Thank you for your quick reply, and I am sorry to disturb you.

Issue #375, opened 2013-05-01
Alternative to multiple keys for one template


magnars:
Good day,

What I am trying to achieve is that I have a template that can either be expanded by typing a complete mnemonic shortcut (say `ascn` for `assert.className`), or to just type a common root to get a list of suggestions (say `as` to get a list of all assertions).

The simplest way of fixing that in my mind would be to allow multiple keys. But as I can tell from the yasnippet source code, it might be an extensive job to make that happen.

So, do you have any suggestions on how to do it? Maybe I should programmatically create the snippets instead, and register two snippets per template?

npostavs:
> The simplest way of fixing that in my mind would be to allow multiple keys. But as I can tell from the yasnippet source code, it might be an extensive job to make that happen.

See also #114.

zw963:
This is a very desirable feature for me too. #342 exist a example for this.

It seem like it still not worked ...

Issue #342, opened 2012-12-04
[Feature request] mulity key binding with one snippet.


zw963:
Define some difference `key value` in one snippet.

e.g.

``` el
# -*- mode: snippet -*-
# name: MiniTest describe
# key: describe
# key: desc
# key: des
# binding: M-c d
# condition: (and (yas-expand-p) (yas-ruby-in-test-file-p))
# --
describe ${1:$(quo)}${1:}${1:$(quo)} do
  `(yas-strip-selected-text)`$0
`(end)`
```

PR #1068, opened 2020-09-02
refactor: mutiple key support

ShuguangSun:
Add support for multiple key for one snippet. For example, the snippet below will be parsed as two snippets with two keys.

```
# name: choose
# key: choose
# key: choosevalue
# --
\\\"\${${1:1}:$$(yas-choose-value '("$0"))}\\\"
```

As this patch modifide the funtion `yas--parse-template` which is a quite core function, I'm not sure whether this patch can be accepted. If it is not acceptable now, I'd like to ask your kindness agreement to me to make a separate package, like the one in
[yasnippet-multiple-key](https://github.com/ShuguangSun/yasnippet-multiple-key). It uses your code and the patch.

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -1615,7 +1615,8 @@ Here's a list of currently recognized directives:
                      (yas--calculate-group file)))
          expand-env
          binding
-         uuid)
+         uuid
+         results)
     (if (re-search-forward "^# --\\s-*\n" nil t)
         (progn (setq template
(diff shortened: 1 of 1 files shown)


ShuguangSun:
Related with: 
https://github.com/joaotavora/yasnippet/issues/375 
https://github.com/joaotavora/yasnippet/issues/114
https://github.com/joaotavora/yasnippet/issues/342

PR #1102, opened 2021-05-11
Prevent yas-new-snippet inserting region twice

mosquito-magnet:
On yas-new-snippet, if yas-wrap-around-region is t and
yas-new-snippet-default has its default value, the selected region is
inserted twice: Once for yas-selected-text and once for $0.

The yas-new-snippet-default snippet is a special usage scenario, where
values of yas-wrap-around-region other than nil don't make much sense,
so override it.


nfedyashev:
+1 I also have this issue. Thanks for working on it @mosquito-magnet

j3wi:
Also having this problem.

PR #1149, opened 2022-10-23
Push modifications to buffer-undo-list, even if unwinding

pestctrl:
I've written a couple of yasnippets for ledger, that involve user input through org-read-date. Something like this:

```
# -*- mode: snippet -*-
# name: Credit Card
# key: cred                                                                       
# --
`(format-time-string "%Y/%m/%d" (org-read-date nil t))` ${0:Charge to Credit Card}
    ${2:Exp}  ${1:\$100}
    Lia:CreditCard
```
However, occasionally I change my mind after expanding the yasnippet, and hit C-g. This completely ruins undo-tree, it's basically unusable. 

The patch I've proposed fixes this, though IDK if there's a more elegant way to fix this. I noticed that if the block of code that modifies the buffer-undo-list still executes after hitting C-g, everything is fine. So, I'm wrapping the whole block that is "inserting" the snippet with an unwind-protect, to ensure that the proper values still get pushed to buffer-undo-list. 

Is this good?

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -4128,39 +4128,40 @@ Returns the newly created snippet."
       (yas--letenv expand-env
         ;; Put a single undo action for the expanded snippet's
         ;; content.
-        (let ((buffer-undo-list t))
-          (goto-char begin)
-          (if (> emacs-major-version 29)
-              ;; Don't use the workaround for CC-mode's cache,
-              ;; since it was presumably a bug in CC-mode, so either
-              ;; it's fixed already, or it should get fixed.
-              (progn
-                (insert content)
-                (narrow-to-region begin (point))
-                (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


pestctrl:
ping

monnier:
Really?  Ping already after waiting what... not even 2 years?  Come on, give us a break!

More seriously: We should use `undo-amalgamate-change-group` instead, but in the mean time your change looks OK (it's going to do the wrong thing if we get an error before we `narrow-to-region`, but that seems much less likely than your case).

But I can't just merge your change because it has conflict with the current code.  Can you fix the conflict?

pestctrl:
Branch has been rebased.

monnier:
Thank you!
And sorry it took so long.

pestctrl:
Yay!

PR #1194, opened 2024-02-29
Undo all-parents memoization in test

gutron:
yas-reload-all does not undo the memoization done in yas--all-parents, which causes the issue-492-and-494 test to fail when run after any tests that expands yas-with-some-interesting-snippet-dirs.

With the memoization enabled, c++-mode is considered a parent of c-mode, causing the issue-492-and-494 test to fail.  With the memoization reset, this does not occur.

### DIFF
--- a/yasnippet-tests.el
+++ b/yasnippet-tests.el
@@ -1438,6 +1438,7 @@ hello ${1:$(when (stringp yas-text) (funcall func yas-text))} foo${1:$$(concat \
   (yas-with-snippet-dirs '((".emacs.d/snippets"
                             ("yas--test-mode")))
     (yas-reload-all)
+    (put 'c-mode 'yas--all-parents nil)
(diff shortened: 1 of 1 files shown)


monnier:
Thanks!
Maybe we should make `yas-reload-all` flush all those `yas--all-parents` caches?

gutron:
`yas--all-parents` caches on the major mode symbols, so clearing those cached values requires knowing which modes are in use. I thought about switching the cache to using a hash table stored in a local variable, but decided that was a bigger change than warranted for fixing that one test.

monnier:
> `yas--all-parents` caches on the major mode symbols, so clearing those
> cached values requires knowing which modes are in use.

It's actually not that bad: `mapatoms` lets us find all the symbols
hence all the modes.
I just pushed a patch which does that.

PR #1196, opened 2024-04-02
(yas--all-parents): delete-dups for fundamental-mode

VitoVan:
Before this commit,

(yas--all-parents 'lisp-mode) would return:

`(lisp-mode lisp-data-mode prog-mode fundamental-mode fundamental-mode)`

which would cause duplicated entries when prompting (if the snippet were for fundamental-mode).

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -848,18 +848,19 @@ which decides on the snippet to expand.")
                             (cons (if (eq mode 'fundamental-mode) ()
                                     (append (cdr ap) '(fundamental-mode)))
                                   extras))))
-                 (cons mode
-                       (yas--merge-ordered-lists
-                        (mapcar #'yas--all-parents
-                                (remq nil
-                                      `(,(or (get mode 'derived-mode-parent)
-                                             ;; Consider `fundamental-mode'
-                                             ;; as ultimate ancestor.
-                                             'fundamental-mode)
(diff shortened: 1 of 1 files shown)


monnier:
> (yas--all-parents #'lisp-mode) would return:
>
> `(lisp-mode lisp-data-mode prog-mode fundamental-mode fundamental-mode)`

Thanks, and sorry for the slow response.

`delete-dups` is right, but it should be applied deeper: the `(get mode
'yas--all-parents)` is basically looking up a cache, so we should
`delete-dups` when we `put` the info in the cache rather than every time
we fetch it from the cache.

Also, I wonder which part of the code you're exercising: is
`derived-mode-all-parents` defined in your Emacs (IOW, are you running
Emacs-30)?  With `derived-mode-all-parents` defined I don't see the
problem, so I think the `delete-dups` should go in the branch where
`(fboundp 'derived-mode-all-parents)` returned nil.

VitoVan:
> so I think the `delete-dups` should go in the branch where `(fboundp 'derived-mode-all-parents)` returned nil.

You are right.

I was being lazy in the first place.

I rebased the commit, and put it in the branch where `(fboundp 'derived-mode-all-parents)` returned nil.

Please let me know if the code format went wrong.

monnier:
Thanks!

PR #1202, opened 2024-08-25
Function to load snippets without confirmation

lanceberge:
Yasnippet by default prompts the user to enter a table, file path to save the snippet, and if they want to save the buffer. 
I found myself almost always applying the defaults, so I wrote a function that saves the snippet, loads it, and quits without any confirmation.

## Testing

M-x `yas-new-snippet`
Fill in a snippet
M-x `yas-load-snippet-buffer-and-close-noconfirm
test that the snippet works

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -2824,6 +2824,23 @@ and `kill-buffer' instead."
         (save-buffer)))
     (quit-window kill)))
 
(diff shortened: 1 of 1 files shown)


Mango0x45:
Instead of `(car (cdr (car ...)))` you can just do `(cadar ...)`

lanceberge:
Cool, added

PR #1207, opened 2025-03-05
Add yas-completion-at-point-function

paaguti:
This adds a simple completion-at-point function for yasnippets. It completes based on the keys that are active for a mode.


paaguti:
If you don't want to add the capf part, consider the fontification issue. I have reported it in emacs-devel. 

[248021a](https://github.com/joaotavora/yasnippet/pull/1207/commits/248021a602a380a69e0937360505bf9de1fb60eb)
was a first try, which didn't completely work

[bcdc746](https://github.com/joaotavora/yasnippet/pull/1207/commits/bcdc746710aacb3d34621f19eb8160486e329ce1) 
reverts that and adds a simple and cleaner solution.

monnier (commented):
[ Thanks for your code, and sorry for the slow response. ]

`yas--getkeys` should probably use `(apply #'nconc (mapcar ...))` or just `mapcan` because your code currently suffers from an unnecessary O(N²) factor.

`yas-completion-props`: Why have this is as a defvar rather than inside the function?  Also, setting `corfu-on-exact-match` seems rather drastic since it will affect all completions rather than only the ones for yasnippet.

`yas-completion-at-point`: AFAICT, `begin` and `end` can never be nil, so we shouldn't needlessly test them.  Also I don't understand the "rationale" in the docstring: it seems ... vacuous?

Finally: how/where do you use it?  If you use it in your init file, maybe add a comment showing what that looks like?

PR #1210, opened 2025-04-02
yasnippet.el(yas-minor-mode-on): autoload

diniamo:
My use-case requires passing that function to a hook, and it's much better to do that without loading the package.

### DIFF
--- a/yasnippet.el
+++ b/yasnippet.el
@@ -974,6 +974,7 @@ per-mode exceptions to the \"global\" activation behaviour, call
 `yas-minor-mode' with a negative argument directily in the major
 mode's hook.") ;; FIXME: Why do we say "Only the global value is used"?
 
+;;;###autoload
 (defun yas-minor-mode-on ()
   "Turn on YASnippet minor mode.
 


monnier:
> My use-case requires passing that function to a hook, and it's
> much better to do that without loading the package.

Could you describe your use case a bit more precisely, especially
explaining why you need to use `yas-minor-mode-on` rather than
`yas-minor-mode`?

diniamo:
Oh wait, I didn't see yas-minor-mode does the same thing with the argument omitted. Why does yas-minor-mode-on exist then?

diniamo:
Yes, I checked that, but yas-minor-mode could be passed to define-globalized-minor-mode instead of yas-minor-mode-on, and it would work the same, while removing the need for that function.

monnier:
> Yes, I checked that, but yas-minor-mode could be passed to
> define-globalized-minor-mode instead of yas-minor-mode-on, and it would work
> the same, while removing the need for that function.

Not quite the same: if we were to pass `yas-minor-mode` instead of
`yas-minor-mode-on` to `define-globalized-minor-mode`, then YAsnippet
would be enabled in *all* buffers, including those where
`yas-dont-activate-functions` says we shouldn't enable YAsnippet.

diniamo:
Ah, I missed that. Well my use case was not having to declare a new function that just calls `(yas-minor-mode 1)`, when passing it to the lsp-mode hook:

```elisp
(add-hook 'lsp-mode-hook 'yas-minor-mode-on)
;; instead of
(add-hook 'lsp-mode-hook (lambda () (yas-minor-mode 1)))
```

But I guess I should just do that then?

monnier:
> Ah, I missed that. Well my use case was not having to declare a new function
> that just calls `(yas-minor-mode 1)`, when passing it to the lsp-mode hook:
>
> ```elisp
> (add-hook 'lsp-mode-hook 'yas-minor-mode-on)
> ;; instead of
> (add-hook 'lsp-mode-hook (lambda () (yas-minor-mode 1)))
> ```
>
> But I guess I should just do that then?

Just use (add-hook 'lsp-mode-hook #'yas-minor-mode), since presumably
you do want YAsnippet in all `lsp-mode` buffers.
`yas-minor-mode-on` is trying to avoid enabling YAsnippet in the
minibuffer, in hidden temp buffers, and things like that where LSP will
presumably not be enabled anyway.

diniamo:
The problem with that is that it will toggle, so in buffers that already have it enabled, it will be disabled. But I can just use a lambda then.

May I ask what #' is for? I'm pretty sure it has something to do woth byte-compilation, but I never found a proper explanation online. From what I've seen it mangles function names in the documentation (e.g., when looking at the value of lsp-mode-hook with `C-h v`), which is annoying.

diniamo:
> What makes you think so?

Well, it says in the function documentation that the function will toggle the mode, if an argument is not passed.

(#') Thanks for the explanation. A little ~~type~~ symbol safety is nice, I'll use it from now.

monnier:
>> What makes you think so?
> Well, it says in the function documentation that the function will toggle the mode, if an argument is not passed.

Ah, I see.  That's the description of the behavior with respect to the
`C-u` prefix argument, when used interactively.  It's a left-over from
many years ago.

The doc should say something like:

```
yas-minor-mode is an autoloaded interactive interpreted-function in
‘yasnippet.el’.

(yas-minor-mode &optional ARG)

YASnippet minor mode.

When YASnippet mode is enabled, ‘yas-expand’, normally bound to
the TAB key, expands snippets of code depending on the major
mode.

This is a minor mode.  If called interactively, toggle the ‘yas minor
mode’ mode.  If the prefix argument is positive, enable the mode, and if
it is zero or negative, disable the mode.

If called from Lisp, toggle the mode if ARG is ‘toggle’.  Enable the
mode if ARG is nil, omitted, or is a positive number.  Disable the mode
if ARG is a negative number.

To check whether the minor mode is enabled in the current buffer,
evaluate the variable ‘yas-minor-mode’.

The mode’s hook is called both when the mode is enabled and when it
is disabled.


Key             Binding

\^c & \^n	yas-new-snippet
\^c & \^s	yas-insert-snippet
\^c & \^v	yas-visit-snippet-file
```

The corresponding change was described here:

```
* Incompatible Lisp Changes in Emacs 24.1

** Passing a nil argument to a minor mode function call now ENABLES
the minor mode unconditionally.  This is so that you can write e.g.

    (add-hook 'text-mode-hook #'foo-mode)

to enable foo-mode in Text mode buffers, removing the need for
'turn-on-foo-mode' style functions.  This affects all mode commands
defined by 'define-minor-mode'.  If called interactively, the mode
command still toggles the minor mode.

```

diniamo:
Oh wow, alright. Thanks again for the help.

monnier:
> Oh wow, alright. Thanks again for the help.

I updated the docstring accordingly, thank you.

PR #1219, opened 2025-10-20
Implement workaround for Emacs semantic highlighting

paaguti:
Add workaround to disable semantic highlighting in snippet mode.
Depending on the theme, the snippet header looks confusing and difficult to read.

This PR overrides the combined capf + semantic PR I submitted yesterday.


monnier (commented):
Can we put the `setq-local` directly inside the definition of `snippet-mode` (and it doesn't even need a `boundp` check, AFAICT)?
Also, it would be good for the comment to mention (maybe as a "FIXME") that the core problem is that `snippet-mode` applies `emacs-lisp-mode` keywords over the whole buffer, including those parts which are plain text rather than ELisp code.

